PR #2, opened 2012-12-05
Changing the empty ISO icon image, removing unused resources

john-peterson:
### Discussion

> [02:55] @Jasper Sonicadvance1, pull request #2 looks fine
> [02:56] @Jasper cuts down on binary size
> [02:56] @Jasper no downside afaict
> [02:57] @Jasper well there's some crazy custom drawing code that doesn't really make sense


john-peterson:
This might have been aesthetically appealing ten years ago. I don't think I will ever bother about this. It seems to only be important in appearance. It has no meaning beyond appearance of various copyrighted images that I would avoid using

I'm looking for engineering that I might be interested in in the future. I think this can be safely forgotten

PR #5, opened 2012-12-05
Adding option to log without file or type

john-peterson:
### Problem

When printing
- hex dumps (f.e. from a Wiimote)
- series of data f.e.
  - formatted Wiimote date
  - Gecko code log

the file and type prefix
- isn't meaningful
- use horisontal space

this example show
- a file and type prefix that's 50 letters
- an indentation offset of 1 letter because the row number is different

```
10:34:860 Src\HW\WiimoteReal\WiimoteReal.cpp:379 N[Wiimote]: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
10:34:861 Src\HW\WiimoteReal\WiimoteReal.cpp:79 N[Wiimote]: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
```

This is a problem because
- it use space i.o.w. leave less room for the log
- it alters the indentation length when there's a difference in
  - number of letters in the row number
  - file path
- it's slower to locate to locate the beginning of log text when
  - there's preceding text that's not different in apperance f.e. color
  - the indentation length is varaible
### Solution

Adding a log without file prefix, that print this instead of the above

```
10:34:860 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
10:34:861 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
```
### Function name

The log name is
- LOG#: exclude file and type
- LOG#1: exclude file

because
- fast to type becuse it's short
- the color number 1, 2, 3, 4, 5 is easier to remember than N, W, E, I, D
### Punctuation

The punctuation is changed from

```
12:35:014 Src\File.cpp:350 N[COMMON]: test 1
```

to

```
12:35:014 Src\File.cpp:350 N[COMMON] test 1
```

because
- that use less horisontal space
- the colon separator isn't meaningful to indicate where the log message start because it can be inferred from the format of the log
### Log selection

> [15:19] <_skidau> yeh, i think the log file name is quite redundant, but other people like it, so don't merge the patch

It doesn't alter the current log f.e. `WARN_LOG`, it adds f.e. a `WARN_LOG_S`

> [15:29] <_skidau> hmm.. i don't like it, sorry :P

Describe the reason for that

> [15:34] <_skidau> the confusion of "why are there two ways to log"

Because
- non-text data is more efficient to print without a file and type prefix because row length and fixed indentation length matters
- some logging is temporary in the sense that it's removed when a problem is solved
- for this logging it's valuable to grant different preferences regarding file and type prefix
### Identify log origin

> [15:45] @Sonicadvance1 And then on that point. If they are left in the source and they accumulate, it can very well turn in to a clusterfuck of "Where are these messages coming from?"

When committing a log without file prefix it's appropriate to prepend it with a log that describe the source of f.e. the following hex dump because it would otherwise have little meaning for the reader
### Value

> [16:15] @Sonicadvance1 This seems to be a very specific change that will be used in just a singular location
> [16:15] @Sonicadvance1 eg, You debugging wiimote dumps

Hex dumps are important for emulation and reverse engineering and the emulator should have
- a log function without a file and type prefix
- a hex dump function that print a C array (i.o.w. the output format of `xxd -i`)

> [16:32] @Sonicadvance1 I think that is called printf

The problem with `printf` is
- it isn't printed in Windows in origin
- it's printed in Windows in a commit in https://github.com/mirror/dolphin-emu/pull/1 but it's not merged

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,30 @@
+language:
+# linux
+- cpp
+
+# osx
+# - objective-c
+
+env:
+# linux
+- build=release
+- build=debug
+- build=cli
+
+install:
+# linux
+- if [[ "$build" != "cli" ]]; then sudo add-apt-repository -y ppa:fransschreuder1/usbpicprog-stable; fi
+- sudo apt-get -qq update
+- sudo apt-get -qq install git cmake gettext libwayland-dev libsdl1.2-dev libxrandr-dev libxext-dev libglew-dev libavcodec-dev libavformat-dev libsoil-dev libsfml-dev libopenal-dev libao-dev libasound2-dev libpulse-dev libportaudio-dev libsoundtouch-dev libbluetooth-dev libreadline-dev libswscale-dev liblzo2-dev
+- if [[ "$build" != "cli" ]]; then sudo apt-get -qq install wx2.9-headers libwxbase2.9-dev libwxgtk2.9-dev libgtk-3-dev; fi
+- export MAKEFLAGS="-j $((`nproc` / 2))"
+
+# osx
+# - brew update
+# - brew install cmake
+# - export MAKEFLAGS="-j 4"
+
+script:
+- if [[ "$build" == "release" ]]; then cmake . && make; fi
+- if [[ "$build" == "debug" ]]; then cmake -DCMAKE_BUILD_TYPE=Debug . && make; fi
+- if [[ "$build" == "cli" ]]; then cmake -DDISABLE_WX=ON . && make; fi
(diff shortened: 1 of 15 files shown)


john-peterson:
## Adding log colors

### Commit message

because
- color makes it faster to identify elements of a log message

PR #6, opened 2012-12-06
Adding AA, AF, video backend to iso properties

john-peterson:
### Discussion

[@kostamar](https://code.google.com/p/dolphin-emu/issues/detail?id=5041)

The Red Steel 2 [OpenGL] backend setting is removed because of this information

@NeoBrainX

> - Will the gfx backend setting work on Linux if Direct3D is specified as the game specific option or will that cause a segfault?

The gfx backend iso property takes the backends from the same list as videoconfig so there is no D3D option in linux

> Please don't make more use of the _connect_macro_ mess, it's awkward and plain wx connections should be used instead

I request that a consensus is demonstrated for this opinion because
- it's used elsewhere
- it's not clear why it's not beneficial

### DIFF
--- a/Source/Core/Common/Src/VideoBackendBase.h
+++ b/Source/Core/Common/Src/VideoBackendBase.h
@@ -98,6 +98,7 @@ public:
 	virtual std::string GetName() = 0;
 
 	virtual void ShowConfig(void*) {}
+	virtual void InitBackendInfo() {}
 
 	virtual void Video_Prepare() = 0;
 	virtual void Video_EnterLoop() = 0;
--- a/Source/Core/Core/Src/BootManager.cpp
+++ b/Source/Core/Core/Src/BootManager.cpp
@@ -29,9 +29,6 @@
 //               Boot.cpp               CBoot::BootUp()
 //                                      CBoot::EmulatedBS2_Wii() / GC() or Load_BS2()
 
-
-// Includes
-// ----------------
 #include <string>
 #include <vector>
 
@@ -47,7 +44,6 @@
 #include "VideoBackendBase.h"
 #include "Movie.h"
 
-
 namespace BootManager
 {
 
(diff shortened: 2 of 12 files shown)


john-peterson:
## wxNotebook drawing bug

### Problem

When selecting
- Right-click → Properties

the 'GameConfig' tab checkboxes
- are drawn for around 0,5 s and then are not drawn (until the mouse cursor move over them)
- non-modal GCPad
- starting a game (that run UpdateGUI)

many controls are not drawn (until the mouse cursor move over them)

### Solution

Drawing all controls on open with

```
m_Notebook->ChangeSelection(1)
m_Notebook->ChangeSelection(0)
```

in InputConfigDialog

```
int selection = m_pad_notebook->GetSelection();
m_pad_notebook->ChangeSelection((selection + 1) % m_pad_notebook->GetPageCount());
m_pad_notebook->ChangeSelection(selection);
```

because

there's no shorter code to force refresh, f.e.

```
Refersh(true)
Update()
m_pad_notebook->GetPage(m_pad_notebook->GetSelection())->Refresh(true);
m_pad_notebook->GetPage(m_pad_notebook->GetSelection())->Update();
```

the problem with this solution is
- this SetFocus to the window which is distracting because it occurs every time UpdateGUI occurs, f.e. on play/pause
- it's more distracting than updating the table manually by moving the cursor over it or chaning tabs

### Discussion

@NeoBrainX

> "// 2.9.2 drawing bug" isn't particularly descriptive for anyone who hasn't ever heard of that bug. There should at least be a link to a description of that bug, or better yet an explanation within the source code comments

It's described in this topic

john-peterson:
This is just custom ISO settings. Typical junior engineering tasks that I should not be bothered with

And of course I came across an interface drawing bug. Don't even like windows. At least not on a desktop. Prefer commands

This would read more from the ISO settings file that could be added manually

But it's just a junior engineering task and a silly window drawing bug. That's probably ancient now. Not anything I need to save desperately

PR #17, opened 2013-05-02
Adding menu option to switch between debugging and default GUI

john-peterson:
### Discussion

@NeoBrainX

> - 3859deb2499f : Don't like the debug switch; should prolly be moved to the View menu and/or disabled for Release builds.

The debug switch menu item is moved to the view menu

It should be in the Release build since that build is faster than DebugFast and has some functionality of the debugging GUI, f.e. code view

> - I don't agree that Release builds need a "restart in debug mode" option. I don't see any need for this option at all, but if anything …

The log states "this makes it easier to switch between debugging and regular mode" because close → start is two actions, restart is one action; easier.

> … it shouldn't be in non-developer builds.

The debugging mode has use in the release build too; additional settings, code inspection.


john-peterson:
This was too contentious to implement. It fell on emotions. I would find it obvious but it was opposed. Interface design usability user interface issue. I don't represent the marketplace or the typical user

PR #21, opened 2013-05-05
Adding revision reader to Wii ISOs

john-peterson:
### Problem

https://github.com/mirror/dolphin-emu/pull/24 require version information
### Discussion

> [03:14] @Sonicadvance1 JPeterson, Pull request #21, We don't care, no.
> 
> [03:27] @Matt_P Sonicadvance1: what is wrong with having a revisiion get method in CVolumeWiiCrypted
> [03:27] @Matt_P to me that is fine, what he does with it may or may not be though
> [03:27] @Sonicadvance1 Matt_P, What's the reason for?
> [03:28] @Sonicadvance1 A revision method is easy sure, but with no clear reason for it. I don't understand why it is needed
> [03:28] @Matt_P Well we could have the revisiion listed in game properties

### DIFF
--- a/Source/Core/DiscIO/Src/VolumeWiiCrypted.cpp
+++ b/Source/Core/DiscIO/Src/VolumeWiiCrypted.cpp
@@ -156,6 +156,18 @@ std::string CVolumeWiiCrypted::GetMakerID() const
 	return makerID;
 }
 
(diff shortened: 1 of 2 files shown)


john-peterson:
It might be extracted with external tools if I even understand what it is. A version of a custom file format?

PR #24, opened 2013-05-06
Considering version for Gecko codes

john-peterson:
### Problem

"Cheats Manager → Gecko Codes" display the codes for all versions, f.e for SF8E01

```
$infinite health [wiiztec]
0480BD8C 00000002

$infinite health [wiiztec]
0480DD8C 00000002
```

This is a problem because
- it's possible to select a code for the wrong version
- the code for the wrong version use space in the list which result in more scrolling

Binary patches should consider the version because the memory location is different between versions
### Solution

The patches should be version aware by having the 0x007 version in the
- ini setting category name f.e. "ActionReplay00"
- filename f.e. GALE0100.ini

F.e. [these](https://github.com/mirror/dolphin-emu/blob/master/Data/User/GameConfig/GALE01.ini) AR codes are for version 0x00
## Versions

There are

three [GALE01](http://www.gametdb.com/Wii/GALE01) versions

two [SF8E01](http://www.gametdb.com/Wii/SF8E01) versions

```
version     apploader date  data at 0x00
0           ?               SF8E01 0x0000
1           2009-12-11      SF8E01 0x0001
```

### DIFF
--- a/Source/Core/Core/Src/ActionReplay.cpp
+++ b/Source/Core/Core/Src/ActionReplay.cpp
@@ -111,7 +111,7 @@ bool CompareValues(const u32 val1, const u32 val2, const int type);
 
 // ----------------------
 // AR Remote Functions
-void LoadCodes(IniFile &ini, bool forceLoad)
+void LoadCodes(IniFile &ini, u8 version, bool forceLoad)
 {
 	// Parses the Action Replay section of a game ini file.
 	if (!SConfig::GetInstance().m_LocalCoreStartupParameter.bEnableCheats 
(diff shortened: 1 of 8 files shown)


john-peterson:
## WiiRD database

### Problem

The WiiRD database

download interface http://geckocodes.org/txt.php doesn't have a version GET variable
- f.e. http://geckocodes.org/index.php?c=SF8E01&page=3 download the same page as http://geckocodes.org/index.php?c=SF8E01 that download codes for both versions

### Solution

Add a version GET variable (f.e. "?v=1")  to http://geckocodes.org/txt.php corresponding to the the version value at 0x007 on the [Wii disc](http://wiibrew.org/wiki/Wii_Disc) because
- that allow download of codes for the correct ISO version
- this is a better solution than parsing the text based on the string "Version 1 Codes" and "Version 2 Codes" because it's simpler code

### Information

This related information can be meaningful

The WiiRD database

browser interface http://geckocodes.org/index.php has a verson variable "page"
- this variable doesn't correspond to the 0x007 version data, f.e. "page=3" corespond to version 0x0001
- f.e. http://geckocodes.org/index.php?c=SF8E01&page=3 is codes for SF8E01 0x0001

### Discussion

@James_Atherton

> What's your endgame? Maybe I can help...

The endgame is described in the topic "Solution"

PR #40, opened 2013-07-03
Added functionality to memory and code view

shua:
The biggest thing I wanted to do was fix hex view and jumping when you try to click around. I also ended up adding scrolling functions to both views (I kept trying to scroll and then remembered I couldn't). I also added middle click to follow branch at current address.

In addition, I cut out some uneeded draw calls and functions, and the 'f:' specifier before every number in float view because I felt it was superfluous.


john-peterson:
## Conflict: Reducing duplication in the code and memory view

> I also ended up adding scrolling functions to both views

Consider basing this commit on the commit with the title "Reducing duplication in the code and memory view" in https://github.com/mirror/dolphin-emu/pull/23 because
- it reduce duplication in the code and memory view
- the scrollbar should be added to CDebugView instead of CCodeView and CMemoryView because it reduce duplication

> and the 'f:' specifier before every number in float view because I felt it was superfluous.

This is done in the mentioned commit because of the same reason

shua:
I just realized, is this your personal clone? I thought it was an official dolphin one. Sorry.

PR #43, opened 2013-10-30
Implement SwapInterval for AGL interface

zakk4223:
Implemented the correct call to the cocoa NSOpenGLContext to set the swap interval, so the VSync video setting works in OSX.


degasus:
Hi, I'm sorry, this isn't the official repo for dolphin, so I guess this pull request just got lost for a long time.

https://github.com/dolphin-emu/dolphin/commit/0eadc2e2a4c4f6d05854c2f2bd2c80ffc9082265

